PR #7, opened 2020-04-19
Add a fuzzing target to verify interoperability with libzstd

Shnatsel:
Add a fuzzing target that verifies that that we can decompress any data compressed by the reference implementation. No failures detected after 200 million executions.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -8,6 +8,9 @@ publish = false
 [package.metadata]
 cargo-fuzz = true
 
+[dependencies]
+zstd = "0.5.1+zstd.1.4.4"
+
 [dependencies.ruzstd]
 path = ".."
 [dependencies.libfuzzer-sys]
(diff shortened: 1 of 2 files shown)


KillingSpark:
Thanks!

Shnatsel:
After 800 million executions this has found no issues. Impressive!

KillingSpark:
Glad to hear!

Is this with a corpus of valid .zst files or is it purely random input?

Shnatsel:
Uh, neither? The fuzzer feeds uncompressed data to the added code in this PR, which compresses it with libzstd and decompresses the result of that with ruzstd. So the fuzzer produces random data, but ruzstd gets a valid zstd as input.

Verifying that the behavior is identical to libzstd given a randomly generated zstd file to decompress is the next thing on my list.

KillingSpark:
Sorry, that was an absolute brainfart. I had my old fuzzing in mind.

Great news though!

PR #9, opened 2020-04-19
Rename "fuzz_target_1" to "decode", add more seeds

Shnatsel:
- Rename "fuzz_target_1" to "decode"
- Move fuzzing corpus to its default location
- Update fuzzing instructions accordingly
- Expand corpus for "decode" target by seeding it with small files first, letting it run for a while and then adding large files
    - Prevents the corpus from blowing up as most of the interesting codepaths are already covered by small files
- Make encoding failure in "interop" target a hard failuire - if encoding fails then the fuzz target is broken
- Add autogenerated fuzzing seeds for "interop" target, obtained by running 800 million executions of it


KillingSpark:
Is it common to include this much data for the fuzzing? I don't know how other projects do this but it seems like a lot.

Shnatsel:
It's a lot of files, but the total size is relatively small: 270kb and 5Mb. That's less than the decode corpus folder (16Mb).

As long as this data is not included in the archive published on crates.io, you're good.

KillingSpark:
Oh I see. I expected this to be quite a lot bigger. Thanks again for your PRs!

PR #11, opened 2020-12-16
A few code improvements

paolobarbolini:
This PR can be reviewed by looking at it commit by commit


KillingSpark:
Hey! Thanks for the improvements. I agree with pretty much all of them, but the commit 095c070.

I think I intentionally cloned and cleared to keep the allocations in the buffer. I guess this is all very hand-wavy without benchmarks and differs from usecase to usecase, but since Vec allocates in 2^x steps that means refilling the buffer results in many allocations while the current method only takes one big allocation for cloning the buffer.

I guess you could use take + Vec::reserve which might make the intent clearer but I think this is fine as it is.

If you could revert that commit I'd be happy to take a final look and merge this :)

(Also do you have any idea why I would do that one byte writing loop you fixed in d1e1017 ? This is really weird to me and I am trying to find any reason at all. Because I hope that past me has reasons to do something that inefficient...)

paolobarbolini:
> I think I intentionally cloned and cleared to keep the allocations in the buffer. I guess this is all very hand-wavy without benchmarks and differs from usecase to usecase, but since Vec allocates in 2^x steps that means refilling the buffer results in many allocations while the current method only takes one big allocation for cloning the buffer.

Cloning a Vec is like allocating a new Vec of the same size and copying everything over. If you really wanted to "keep" (we are doing a new allocation anyway, you can't have two `Vec`s sharing the same underlying memory) the capacity of `self.buffer` we could do something like this, which is like cloning but without having to copy all of the data over.

```rust
pub fn drain(&mut self) -> Vec<u8> {
    self.hash.write(&self.buffer);
    std::mem::replace(&mut self.buffer, Vec::with_capacity(self.buffer.capacity()))
}
```

paolobarbolini:
> (Also do you have any idea why I would do that one byte writing loop you fixed in d1e1017 ? This is really weird to me and I am trying to find any reason at all. Because I hope that past me had reasons to do something that inefficient...)

I think you copied and pasted it from here https://github.com/KillingSpark/zstd-rs/commit/49d3c4f786f89a6fa61c309dfa7e5c02a452ee32#diff-f48fec557cf88d4528cd057914ba24833bb755a5d7fc58426a18f0f032dfbe2aR61-R65

KillingSpark:
> Since the clone was just to create a new vec with the same capacity, I changed it in 84053d8 to do just that instead of also cloning all of the data just to clear it away in the next line.

Yep that is definitely better / clearer than the current code. 

Thanks for the changes. I'll make a new release with these changes soon :)

PR #12, opened 2020-12-19
clippy: fix warnings

paolobarbolini:


KillingSpark:
I'll have to check that none of the casts were actually important but most seem to be redundant.

Could you tell me how you called clippy to find these? It does not produce warning for them with just `cargo clippy`.

paolobarbolini:
I denied those warnings, which are allowed by default, by adding that line in lib.rs which you can see in the PR

KillingSpark:
Could you tell me how you called clippy to find these? It does not produce warning for them with just `cargo clippy`.

KillingSpark:
And thanks for the contribution of course!

KillingSpark:
> I denied those warnings, which are allowed by default, by adding that line in lib.rs which you can see in the PR

I see. I thought clippy told you to do that and I was a bit confused :)

PR #16, opened 2022-04-03
Fix latest clippy warnings

paolobarbolini:


KillingSpark:
Thanks for tidying up. Keeping clippy happy is constant upkeep work :)

PR #17, opened 2022-04-03
Remove hot calls to `memmove` by replacing `Vec<u8>` with `VecDeque<u8>`

paolobarbolini:
This removes the need to constantly `drain`, which is one of the hottest functions on the flamegraph.

From a quick benchmark on a large file decoding goes from about 10.5s to 3.5s. As a comparison zstd 1.4.8 takes 1.2s. We're getting closer :eyes:

## Before

flamegraph

## After

flamegraph

### DIFF
--- a/src/decoding/decodebuffer.rs
+++ b/src/decoding/decodebuffer.rs
@@ -17,12 +17,7 @@ pub struct Decodebuffer {
 impl io::Read for Decodebuffer {
     fn read(&mut self, target: &mut [u8]) -> io::Result<usize> {
         let max_amount = self.can_drain_to_window_size().unwrap_or(0);
-
-        let amount = if max_amount > target.len() {
-            target.len()
-        } else {
-            max_amount
-        };
+        let amount = max_amount.min(target.len());
 
         let mut written = 0;
         self.drain_to(amount, |buf| {
@@ -224,11 +219,7 @@ impl Decodebuffer {
     }
 
     pub fn read_all(&mut self, target: &mut [u8]) -> io::Result<usize> {
-        let amount = if self.buffer.len() > target.len() {
-            target.len()
-        } else {
-            self.buffer.len()
-        };
+        let amount = self.buffer.len().min(target.len());
 
         let mut written = 0;
         self.drain_to(amount, |buf| {
--- a/src/decoding/decodebuffer.rs
+++ b/src/decoding/decodebuffer.rs
@@ -1,9 +1,11 @@
+use std::collections::VecDeque;
 use std::hash::Hasher;
-use std::io;
+use std::{io, mem};
+
 use twox_hash::XxHash64;
 
 pub struct Decodebuffer {
-    buffer: Vec<u8>,
+    buffer: VecDeque<u8>,
     pub dict_content: Vec<u8>,
 
     pub window_size: usize,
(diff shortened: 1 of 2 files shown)


KillingSpark:
Thats great! I'll take a look later today and merge this

I wonder how I missed that. I seem to remember that the code taking the most time was the bitreader stuff

paolobarbolini:
> I wonder how I missed that. I seem to remember that the code taking the most time was the bitreader stuff

I guess that might have been caused by me trying ruzstd on a big file compressed with `zstd --ultra -22` :sweat_smile:

KillingSpark:
The TODO around the `extend_from_within` needs to be resolved. This is a massive performance boost for files that have long regions that can be copied from the existing buffer. It should be possible to do, it will likely require some unsafe though. I had an unsafe implementation for this before `Vec::extend_from_within` existed, and I am willing to have some unsafe in there for this big boost in performance.

If you don't want to do that I can merge this and do it myself when I find the time. The code is still correct, it is just being pretty inefficient in these cases.

Otherwise just a few nitpicks,

1. `drain_to` should probably call `f` before the writes to the hash. Because the bytes only get drained when the calls succeed. This might result in the data being put into the hash multiple times. There might be legitimate reasons to call this again even if `f` fails for some reason. I apparently failed to see this when I wrote this code, but I think it would be better this way. Same applies to all places where data is drained. It should first be given to the fallible functions and only if they succeed the data should be given to the hasher.
 
2. `drain_to` should probably not call `f` if n2 is 0. I know it is just a slice with no elements but I think it would "nice" to not do that. And the check is easy enough to do
 
So (imo) it should be
 ```rust
    f(slice1)?;
    hash(slice1);
    if !slice2.is_empty() {
        f(slice2)?;
        hash(slice2);
    }
```
3. drain_to_writer should probably be implemented using `drain_to` too, like `drain_to_window_size_writer` right? Maybe I am just overlooking something here?

**Comment by paolobarbolini** [comment]
Done. Unfortunately `VecDeque` doesn't provide the unsafe APIs to reproduce the equivalent of extend_from_within so I go though an intermediate buffer, which brings a small performance boost for my big file too.

**Comment by KillingSpark** [comment]
Ah right, I remembered why VecDequeue cant implement this nicely. I might play around with a custom buffer structure for this. We don't need all the features a VecDequeue provides. But that is for another day, I'd call this good enough. I too am not sure what a good size for that buffer is, but 4kb seems fine.

One last thing just came to my mind, regarding the functions in drain_to. Right now it is like this:
```rust
   f(slice1)?;
   hash(slice1);
   if !slice2.is_empty() {
       f(slice2)?;
       hash(slice2);
   }
   drain(amount);
```

When we commit slice1 to the hasher, we should also drain that. So when f(slice2) returns an error, we want to still call `drain(0..n1)`. So something like this maybe?

```rust
    f(slice1)?;
    hash(slice1);

    if n2 != 0 {
        let res = f(slice2);
        let drain_amount;
        if res.is_ok() {
            hash(slice2);
            drain_amount = n1+n2;
        } else {
            drain_amount = n1;
        }
        drain(drain_amount)
        res
    } else {
        drain(n1);
        Ok(())
    }
```

Sorry for doing another roundtrip but I think it's good to fix this in this PR too. The drain_to is a very good idea, and having it do the right thing when it's merged in would be nice :)

paolobarbolini:
I'll take a look at your proposal later in the day, this PR just went by in my notifications and it could be very cool to leverage this implementation in the future if it gets optimized more

https://github.com/rust-lang/rust/pull/95632

KillingSpark:
Very nice. Thank you!

PR #19, opened 2022-04-13
Ringbuffer improvements

paolobarbolini:
Let's see if these help.

Inspired from how the std does it


KillingSpark:
This changed slightly from x5.5 to x5.48 which might just be because of whatever. But the improvements around the allocation are still worth it.

paolobarbolini:
:partying_face:. I'll send more improvements in the following days

PR #21, opened 2022-04-25
Update Dependencies

Titaniumtown:
Just noticed a few things were out of date. I updated these dependencies, everything builds correctly and tests still pass 👍🏻

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,12 +11,12 @@ exclude = ["decodecorpus_files/*", "dict_tests/*", "fuzz_decodecorpus/*"]
 readme = "Readme.md"
 
 [dependencies]
-byteorder = "1.3.4"
-twox-hash = { version = "1.6.0", default-features = false }
+byteorder = "1.4.3"
+twox-hash = { version = "1.6.2", default-features = false }
 
 [dev-dependencies]
 criterion = "0.3"
-rand = "0.7.3"
+rand = "0.8.5"
 
 [[bench]]
 name = "reversedbitreader_bench"


Cargo.toml
```diff
@@ -11,13 +11,13 @@ exclude = ["decodecorpus_files/*", "dict_tests/*", "fuzz_decodecorpus/*"]
 readme = "Readme.md"
 
 [dependencies]
-byteorder = "1.3.4"
-twox-hash = { version = "1.6.0", default-features = false }
+byteorder = "1.4.3"
+twox-hash = { version = "1.6.2", default-features = false }
```
paolobarbolini, line 15: What's the reason for bumping those? `cargo update` is already able to update them.

Titaniumtown: Oh really? 😕

paolobarbolini: Let's see what @KillingSpark thinks

KillingSpark:
> What's the reason for bumping those? cargo update is already able to update them.

@paolobarbolini I am confused, wouldn't the changes made by `cargo update` still need to be commited and pushed?

@Titaniumtown does running cargo update change anything else that's worth commiting?

paolobarbolini:
> @paolobarbolini I am confused, wouldn't the changes made by `cargo update` still need to be commited and pushed?

What I meant is we could hold off from bumping `byteorder` and `twox-hash` because `cargo update` is already able to update them for downstream projects, so unless those older versions are broken or insecure there's no need in raising the minimum version that cargo is allowed to choose from

Titaniumtown:
> @Titaniumtown does running cargo update change anything else that's worth commiting?

@KillingSpark Nope! Just doubled checked.

KillingSpark:
> What I meant is we could hold off from bumping `byteorder` and `twox-hash` because `cargo update` is already able to update them for downstream projects, so unless those older versions are broken or insecure there's no need to raise the minimum version that cargo is allowed to choose

But doesn't the cargo.toml that  is in ruzstd right now pin the versions for downstream projects too? I thought we'd need something like `byteorder >= "1.4.3"` instead of `byteorder = "1.4.3"` for downstream to use any compatible version of byteorder

paolobarbolini:
Nope. `1.3.4` actually means `>=1.3.4, <2.0.0`

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

paolobarbolini:
Specifying `1.6` is the same as doing `1.6.0`, while `1.3.4` allows `a.b.c` where `a == 1 && ((b == 3 && c >= 4) || b >= 4)`

KillingSpark:
Yeah, thats what I meant. The versions right now are not really chosen because 1.4.0 was explicitly bad, it was just whatever was the current newest version back then. So going to 1.6(.0) and 1.4(.0) would give users the most choice.

So @Titaniumtown I think that's how it should be. Could you make these changes? 

Always nice to learn something new :)

paolobarbolini:
FYI If you're on nightly you can do `cargo update -Z minimal-versions`, which downgrades your Cargo.lock to the lowest allowed by your direct and indirect dependencies Cargo.toml

You can then run cargo-audit or cargo check to see if the minimal versions are secure and working or are too low

Titaniumtown:
Done

Titaniumtown:
This is confusing. After this PR was merged, it seems `tests::dict_test::test_dict_decoding` now fails. I don't know why.

KillingSpark:
Nope, that came with another commit afaict. I will investigate this, don't worry about it, if you don't want to :)

Titaniumtown:
Ohhh ok. I was very confused lol. Thanks for the info @KillingSpark!

PR #22, opened 2022-04-25
Fixed small typos in comments and Readme

Riuqlav:
Fixed small typos in Comments and Readme


KillingSpark:
Thanks! I don't imagine you actually proof-read all of this. Do you mind sharing what tool you used to check this? I might let it run over my other projects too :)

Riuqlav:
I actually went through everything, I only use the extension code spell
checker, but without it the vscode alone does a little suggestions, you
just go and fix it

KillingSpark:
Oh damn. Then extra thanks for the work!

KillingSpark:
Hehe that actually broke a test that relied on that typo existing. Don't worry its already fixed just wanted to let you know :)

PR #27, opened 2022-06-11
ringbuffer: replace `*mut u8` with `NonNull<u8>`

paolobarbolini:
Mainly done to remove the possibility of `buf` being null, which may cause some bad stuff to happen if we get a reference out of it. It may also help with some minor code optimizations.

### DIFF
--- a/src/decoding/ringbuffer.rs
+++ b/src/decoding/ringbuffer.rs
@@ -1,7 +1,7 @@
-use std::{alloc::Layout, ptr::slice_from_raw_parts};
+use std::{alloc::Layout, ptr::NonNull, slice};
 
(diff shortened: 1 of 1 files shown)


KillingSpark:
I just checked what rusts RawVec is using. They have a `std::ptr::Unique` which is probably what we should be using too.

paolobarbolini:
> I just checked what rusts RawVec is using. They have a `std::ptr::Unique` which is probably what we should be using too.

Unfortunately it's not stable yet. https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/std/ptr/struct.Unique.html

KillingSpark:
Oh, well, then this is good :)

PR #30, opened 2023-03-05
Derive debug and clone for most public facing types

FlareFlo:
Still requires a version bump

### DIFF
--- a/src/blocks/sequence_section.rs
+++ b/src/blocks/sequence_section.rs
@@ -1,9 +1,10 @@
+#[derive(Clone, Debug)]
 pub struct SequencesHeader {
     pub num_sequences: u32,
     pub modes: Option<CompressionModes>,
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug)]
 pub struct Sequence {
     pub ll: u32,
     pub ml: u32,
@@ -16,7 +17,7 @@ impl std::fmt::Display for Sequence {
     }
 }
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub struct CompressionModes(u8);
 pub enum ModeType {
     Predefined,
--- a/src/decoding/decodebuffer.rs
+++ b/src/decoding/decodebuffer.rs
@@ -5,6 +5,7 @@ use twox_hash::XxHash64;
 
 use super::ringbuffer::RingBuffer;
 
+#[derive(Debug, Clone)]
 pub struct Decodebuffer {
     buffer: RingBuffer,
     pub dict_content: Vec<u8>,
--- a/src/decoding/dictionary.rs
+++ b/src/decoding/dictionary.rs
@@ -5,6 +5,7 @@ use crate::decoding::scratch::HuffmanScratch;
 use crate::fse::FSETableError;
 use crate::huff0::HuffmanTableError;
 
+#[derive(Debug, Clone)]
 pub struct Dictionary {
     pub id: u32,
     pub fse: FSEScratch,
--- a/src/decoding/ringbuffer.rs
+++ b/src/decoding/ringbuffer.rs
@@ -1,5 +1,6 @@
 use std::{alloc::Layout, ptr::NonNull, slice};
 
+#[derive(Debug, Clone)]
 pub struct RingBuffer {
     buf: NonNull<u8>,
     cap: usize,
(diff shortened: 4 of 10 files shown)


KillingSpark:
Hi thanks for the  PR!

I'll have a closer look tomorrow, but it seems you added a derive clone for the ringbuffer which holds a raw pointer. That is certainly not correct with respect to dropping. It would lead to a double free.

FlareFlo:
> Hi thanks for the PR!
> 
> I'll have a closer look tomorrow, but it seems you added a derive clone for the ringbuffer which holds a raw pointer. That is certainly not correct with respect to dropping. It would lead to a double free.

I did not dig deeper to check where unsafe is used in said aspect, however, my intent specifically was to be able to clone the `FrameDecoder`, in the context of sharing a dictionary across multiple files that need parsing.

KillingSpark:
> my intent specifically was to be able to clone the FrameDecoder, in the context of sharing a dictionary across multiple files that need parsing.

Wouldn't it be better to provide a way of only parsing the dict once and adding it to the separate framedecoders?

So it would be similar to [init_with_dict](https://docs.rs/ruzstd/0.3.0/ruzstd/frame_decoder/struct.FrameDecoder.html#method.init_with_dict) but instead of taking a raw dict as a `&[u8]` it would just take a `Dictionary`?

* init_with_dict -> init_with_decoded_dict
* reset_with_dict -> reset_with_decoded_dict
* add_dict -> add_decoded_dict

FlareFlo:
I considered the possibility of recycling the entire Decoder, but found no documented way to do so?

KillingSpark:
The `reset` / `init` function should provide this

FlareFlo:
> The `reset` / `init` function should provide this

I will attempt that and see how far it gets me

FlareFlo:
> The `reset` / `init` function should provide this

Attempting to use this, I run into the problem, that initializing the `StreamingDecoder<&[u8]>` as such: 
`StreamingDecoder::new_with_decoder(Default::default(), frame_decoder).unwrap()`, fails with:
`ReadFrameHeaderError(MagicNumberReadError(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })`.
I only really initialize its buffer during the `init` call, after which I read to the end, and loop to the next file.

FlareFlo:
> That error is coming from a call to `io::Reader`
> 
> `r.read_exact(&mut buf).map_err(err::MagicNumberReadError)?;`
> 
> There is probably something wrong with the file, the magic number are the first 4 bytes in a file.

It panics during `new_with_decoder` with an empty array

KillingSpark:
Yes it tries to read the frame description from that empty array and fails. You need to provide it with actual data to initialize the decoder from

FlareFlo:
> Yes it tries to read the frame description from that empty array and fails. You need to provide it with actual data to initialize the decoder from
> 
> What did you expect to happen if you called it with an empty array?
> 
> It would probably help if you could show the code you are trying to get to work :)

Well, how do I initialize it without having data to decode yet? My goal is to create the Decoder with a shared dictionary before entering the hot-loop that processes the files one by one.

KillingSpark:
Something along these lines should do the trick right?

```rust
let mut frame_decoder = FrameDecoder::new();
frame_decoder.add_dict(...);

for file in files {
    frame_decoder = decode_one_file(frame_decoder, file);
}

fn decode_one_file(frame_decoder: FrameDecoder, file: &[u8]) -> FrameDecoder { 
    let mut decoder = StreamingDecoder::new_with_decoder(file_content, frame_decoder);
    // use the decoder
    return decoder.inner();
}
```

FlareFlo:
> Something along these lines should do the trick right?
> 
> ```rust
> let mut frame_decoder = FrameDecoder::new();
> frame_decoder.add_dict(...);
> 
> for file in files {
>     frame_decoder = decode_one_file(frame_decoder, file);
> }
> 
> fn decode_one_file(frame_decoder: FrameDecoder, file: &[u8]) -> FrameDecoder { 
>     let mut decoder = StreamingDecoder::new_with_decoder(file_content, frame_decoder);
>     // use the decoder
>     return decoder.inner();
> }
> ```
> 
> Edit: but really the streaming decoder should take a `AsMut<FrameDecoder>` so it can use a `&mut FrameDecoder` as well as a `FrameDecoder` to make reusing the `FrameDecoder` easier

My issue with that approach was, that constructing and parsing the dictionary over and over got very expensive. For scale: I'm running through 15k files, less than 500 bytes each, with a dictionary of up to 2 MB. Looking at allocation and performance profiling, building the decoders takes a pretty significant amount of time compared to actually processing content.

KillingSpark:
> My issue with that approach was, that constructing and parsing the dictionary over and over got very expensive.

But with that approach you only need to do it once? Or did I misunderstand something?

FlareFlo:
I managed to work around it by passing ownership up again.

PR #31, opened 2023-03-07
Decode multiple frames if present in each file

phord:
A file or stream may consist of more than one frame. Do not stop after the first frame when decoding in zstd. Continue reading until we reach the end of the file or we get an error.

Move percent-done output into do_something() to avoid duplicating code. Calculate the percentage as the ratio of bytes-in instead of bytes-out since we won't know the total bytes-out until we process the last frame.

Report the number of frames as well as the number of bytes at the end. Report total valid checksums at the end instead of per-frame.

### DIFF
--- a/src/bin/zstd.rs
+++ b/src/bin/zstd.rs
@@ -1,8 +1,8 @@
 extern crate ruzstd;
 use std::fs::File;
 use std::io::Read;
-use std::io::Write;
 use std::io::Seek;
+use std::io::Write;
 
 struct StateTracker {
     bytes_used: u64,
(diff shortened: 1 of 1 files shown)


KillingSpark:
Thanks for the PR! 

The idea was to just reuse the `FrameDecoder` if the input file has more than one frame. I think this should be a separate type, something like a `MultiFrameDecoder` that wraps a `FrameDecoder` that calls `reset`/`init` again on the input as long as there are still bytes to process.

phord:
fwiw - I assume it would be trivial-ish to decode a multi-frame file in parallel, assuming the memory is available.

KillingSpark:
Do you know if these multi-frame files are common? I assumed that was more or less a technicality to allow streaming multiple files conveniently over a socket/pipe

phord:
I don't know how common they are in practice. At my day job, we have a few million files compressed with multiple frames (usually over 100 frames each), which is how I noticed it wasn't supported in your cli.

It's a feature commonly available in gzip and other compression formats, and I assume the idea was adopted from there.  It's very useful for parallel compression/decompression as well as quasi-random access into a compressed file.

> RFC 8478                    application/zstd                October 2018
> 
> 3.1.  Frames
> 
>    Zstandard compressed data is made up of one or more frames.  Each
>    frame is independent and can be decompressed independently of other
>    frames.  The decompressed content of multiple concatenated frames is
>    the concatenation of each frame's decompressed content.

Multiframe files are not directly supported by the original zstd cli yet, but there's a [feature request to add it](https://github.com/facebook/zstd/issues/2121). There's also a ["seekable format" spec](https://github.com/facebook/zstd/blob/dev/contrib/seekable_format/zstd_seekable_compression_format.md) that relies on this decode ability to create compliant zstd files which are also more seekable.

It is trivial to create a multiframe file by simply concatenating multiple single-frame files.

    git/opc/zstd-rs» cat LICENSE | zstd > test.zst
    git/opc/zstd-rs» cat test.zst test.zst test.zst test.zst > 4test.zst
    git/opc/zstd-rs» target/release/zstd -d -c 4test.zst > /dev/null
    File: 4test.zst
    100 % done
    Decoded frames: 4  bytes: 4300
    4 of 4 checksums are ok!

btw - I haven't checked to see if your code supports skippable frames (by simply emitting no data, perhaps).  This is also required to be able to read seekable zstd files.

KillingSpark:
Thanks for the PR! I have opened an issue for the skippable frames. They must be supported at least in the way of actually skipping them.

PR #32, opened 2023-03-07
Fix missing variable in error output

phord:
I noticed because the compiler complains about it.


KillingSpark:
Thanks! I wonder why this hasn't caused issues until now...

PR #36, opened 2023-03-12
zstd: Use range slices instead of resizing vectors

phord:
There should be no appreciable difference in runtime, but it seems friendlier and simpler to avoid specialty-mutations for the vector, especially since the function accepts a slice anyway.

--

This is so obvious and more idiomatic, I am curious if it was not done this way originally for a reason.


KillingSpark:
> This is so obvious and more idiomatic, I am curious if it was not done this way originally for a reason.

I'd guess this is because it is code from the very beginning of my rust journey :)

PR #43, opened 2023-07-06
Fix code example in README

Shnatsel:
It will now compile, given the right imports.

It would be best to provide the full code as an example too.


KillingSpark:
Thanks! I always debate having code examples in the readme because there is no tooling (I know of) that would alert me to such breakages. OTOH I think the Readme should contain at least some examples...

Issue #47, opened 2023-09-12
Crate isn't really no-std compatible


tomaka:
https://github.com/KillingSpark/zstd-rs/pull/39 supposedly added support for no-std.

However I strongly disagree with this point:

> Uses of std::error::Error are conditionally replaced with core::error::Error. This is currently unstable, so nightly is required for no_std usage for the time being. I believe it is still not possible to build a no_std binary without nightly, so this should be an acceptable requirement.

You can't really consider a crate to be "no-std compatible" if it requires nightly.
And let's face it: [at the rhythm where it's going](https://github.com/rust-lang/rust/issues/103765) (the first attempts at moving this trait were in 2021), it's going to take a very long time for `core::error::Error` to be stabilized.

My suggestion would be to not derive the `Error` trait at all, but only `Debug` and `Display` and possible `Clone`. This can be done by ditching the `thiserror` crate and using a different one.
After all, the `Error` trait isn't very useful. All it does is potentially provide a stack trace. It's not because it's named "error" that types that represent errors always have to derive it.

KillingSpark:
I am by no means an expert in the no-std field and the PR seemed to satisfy the person that wanted to use the crate in a no_std environment

> I believe it is still not possible to build a no_std binary without nightly, so this should be an acceptable requirement.

Is this no longer true? If it isn't it changes the trade-offs. If it is still true I don't see the point you are trying to make here: `You can't really consider a crate to be "no-std compatible" if it requires nightly.`

> My suggestion would be to not derive the Error trait at all, but only Debug and Display and possible Clone. This can be done by ditching the thiserror crate and using a different one.

`thiserror` is not only useful for deriving the error crate but it makes the display messages really comfortable to format within the `#derive` tags. 

> After all, the Error trait isn't very useful. All it does is potentially provide a stack trace.

The stacktrace is a very valuable source of information if anything goes wrong and I have to debug it. I guess not having a stacktrace in no_std is fine, but completly ditching it seems weird to me, but it is maybe be worth it?

Would you have a suggestion for a better crate to make display messages as easy to create as with thiserror that is no_std compatible?

tomaka:
> Is this no longer true?

I think that since https://github.com/rust-lang/rust/issues/66741 it is possible to create no-std binaries. This issue was the last blocker if I'm not mistaken.
Regardless, there are many use cases that do not involve compiling a binary, for example embedding a Rust library within another program.

[...]

PR #50, opened 2023-10-14
Remove thiserror in favor of derive_more

KillingSpark:
This switches to derive_more so we get

1. Nice formatting of the errors (both for no_std and std)
2. Get to be no_std on stable instead of only nightly
3. Still have the Error implementation if the std feature is enabled

Fixes: #47

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ rand = {version = "0.8.5", features = ["small_rng"]}
 
 [features]
 default = ["std"]
-std = []
+std = ["derive_more/error"]
 
 [[bench]]
 name = "reversedbitreader_bench"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ruzstd"
-version = "0.4.0"
+version = "0.5.0"
 authors = ["Moritz Borcherding <moritz.borcherding@web.de>"]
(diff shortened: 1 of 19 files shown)


tamird:
LGTM. Is this ready to land?

tamird:
You might want to change CI to test against stable rather than nightly: https://github.com/KillingSpark/zstd-rs/blob/e15eb1e568e0339c25db609e4a093ff1b6a90e6d/.github/workflows/ci.yml#L78-L140

tamird:
Dang, `derive_more` also depends on `syn@1`. Severing that dependency edge was my motivation for #48.

KillingSpark:
> Dang, derive_more also depends on syn@1

Yep but they are having beta releases for a 1.0 which makes ma hopeful that this will soon be rectified

I'm pretty busy right now, I'll try and merge this next week

KillingSpark:
Okay so I think the way I'm going to handle this is merging this, and bumping derive_more to 1.0 once they release it. 

This will allow all three goals above to be reached now and gives a clear-ish roadmap (wait for derive_more 1.0 release) towards not compiling syn 1.0 AND syn 2.0 in the future.

tamird:
Cool, thanks. Will you cut a new release with this change?

KillingSpark:
Today @tamird ;)

PR #51, opened 2023-10-26
Replace xxhash with ahash

tamird:
ahash is much more widely used and claims to be a good bit faster.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ exclude = ["decodecorpus_files/*", "dict_tests/*", "fuzz_decodecorpus/*"]
 readme = "Readme.md"
 
 [dependencies]
+ahash = { version = "0.8", default-features = false }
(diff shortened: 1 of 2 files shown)


KillingSpark:
I'm sorry if I am just confused, but ahash is different from xxhash right?

The xxhash algorithm is required by the zstd spec for the checksums, this wasn't my choice

tamird:
Yeah, I see that now. I'm trying to move xxhash to a dev dependency, would you accept that change?

tamird:
Oof, that is seeming pretty difficult given how much `.reset()` is going on.

Well, it'd be good to make xxhash a dev-dependency, but I guess I won't be doing that here.

KillingSpark:
Sorry I should have clarified, xxhash is an *integral* part of the zsdt format. The checksums that are part of the frame definition are calculated that way. You need this if you want to make sure you decoded the correct data, which is probably what you want to do in many cases

PR #53, opened 2023-10-26
Make twox-hash an optional dependency

### LINKED ISSUE #52: Move twox-hash dependency to dev-dependencies
Looks like it is only used for corpus testing. It'd be good to make the hasher optional - either through a type parameter or using a dyn box. Both options will require a good bit of plumbing.

tamird:
Add a new feature "hash" which pulls in twox-hash and hashes during
decoding. Add "hash" to default features.

Fixes #52.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,15 +12,16 @@ readme = "Readme.md"
 
 [dependencies]
 byteorder = { version = "1.5", default-features = false }
-twox-hash = { version = "1.6", default-features = false }
+twox-hash = { version = "1.6", default-features = false, optional = true }
 derive_more = { version = "0.99", default-features = false, features = ["display", "from"] }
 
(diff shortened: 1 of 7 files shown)


KillingSpark:
Looks good! Any reason to change the Hasher initialization from the explicit seed to Default::default? I think it's better to have the explicit thing to make sure we stay spec compliant even if the default seed changes

tamird:
> Looks good! Any reason to change the Hasher initialization from the explicit seed to Default::default? I think it's better to have the explicit thing to make sure we stay spec compliant even if the default seed changes

The only reason was to reduce the number of references to the twox_hash type. I'm not at a computer for a bit but feel free to amend!

KillingSpark:
Thanks! Did the amend and merged :)

PR #58, opened 2024-04-01
don't return errors on too large requests on a reversed bitreader

KillingSpark:
Removing the checks for too large requests (> 56 bits) on a reversed bitreader allows for a whole bunch of error handling in hot loops to be removed which leads to better performance.

Doing this allows more weird values to float around requiring more careful handling of the returned values in many places, more fuzzing is definitely necessary.


KillingSpark:
This should include a change that checks the decoded FSE tables so that the symbols do match the allowed ranges for literal length, match length and offset which would avoid weird values appearing and would also detect corrupted files earlier and more consistently.

PR #59, opened 2024-04-01
Document how to deal with multi-segment streams

### LINKED ISSUE #57: Not decompressing the full stream
Inside [this](https://github.com/KillingSpark/zstd-rs/files/14737107/l0.zip) zip archive there is a zstandard compressed file, produced by blender. (I had to compress it into a zip archive, because github does not allow uploading random binaries)
```shell
$ file l0.blend 
Zstandard compressed data (v0.8+), Dictionary ID: None
$ zstdcat -l l0.blend
Frames  Skips  Compressed  Uncompressed  Ratio  Check  Filename
    18      1   102.38 KB     902.79 KB  8.818   None  l0.blend
```
`zstd` crate, when used to decompress, produces a buffer of length `924460`, while ruzstd only decompresses `66804` bytes

The code to repro:
```rust
let mut decoded = ruzstd::StreamingDecoder::new(File::open(fname).unwrap()).unwrap();
println!("ruzstd {}", decoded.read_to_end(&mut vec![]).unwrap());

let mut decoded = zstd::Decoder::new(File::open(fname).unwrap()).unwrap();
println!("zstd {}", decoded.read_to_end(&mut vec![]).unwrap());
```
versions used
```toml
ruzstd = "=0.6.0"
zstd = "=0.13.0"
```

Am I using it wrong or is there a bug?

Thank you for maintaining this! :heart:

Sorseg:
Fixes #57

Thank you for maintaining this!

### DIFF
--- a/src/bin/zstd.rs
+++ b/src/bin/zstd.rs
@@ -62,10 +62,10 @@ fn main() {
 
         while tracker.file_pos < tracker.file_size {
             match frame_dec.reset(&mut f) {
-                Err(FrameDecoderError::ReadFrameHeaderError(ReadFrameHeaderError::SkipFrame(
-                    magic_num,
-                    skip_size,
-                ))) => {
+                Err(FrameDecoderError::ReadFrameHeaderError(ReadFrameHeaderError::SkipFrame {
+                    magic_number: magic_num,
+                    length: skip_size,
+                })) => {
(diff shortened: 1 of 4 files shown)


KillingSpark:
Thanks for the contribution!

Issue #55, opened 2024-01-25
follow up derive_more release 1.0


yangfengzzz:
please follow derive_more to upgrade 1.0 version. 

https://github.com/JelteF/derive_more/issues/259

Because it will use `syn` 2.0 and a lot other package will use it too. Now in our project, there are two version of `syn`.

KillingSpark:
Yep it's on my todo list, but if I'm not mistaken 1.0 is not yet released. Right?

xd009642:
Another potential option would be to remove derive_more and add in the manual implementations. I know it's a bit of faff but from `cargo build --timings` I can see syn takes up ~25s of my clean build time.  Provided  there's not too much churn in those types it might be a nicer solution provided someone is willing to go over and write all the boilerplate (I'm willing to PR this if @KillingSpark is okay with the change).

KillingSpark:
I'd accept that, the thought crossed my mind too. I just didn't want to spend the time to actually implement it. If you have the motivation to do this, go for it :)

mstange:
> provided someone is willing to go over and write all the boilerplate

Here's a partial attempt I made a few weeks ago: https://github.com/mstange/zstd-rs/commits/displaydoc-instead-of-derive_more/

It uses `displaydoc::Display` and manually implements `From` and `Error`. I initially wanted to use `thiserror` but didn't want to break `no_std` support.

Then I discovered that `thiserror_no_std` exists and that you can do this instead:

```rust
#[cfg_attr(feature = "std", derive(thiserror::Error))]
#[cfg_attr(not(feature = "std"), derive(thiserror_no_std::Error))]
```

So I think using `thiserror` + `thiserror_no_std` would be even better.

KillingSpark:
We actually switched from thiserror_no_std to derive_more. I don't remember why and am on mobile so looking it up is a bit difficult. But I think if this issue is tackled again I would want it to just stay dealt with. Which means either derive_more finally releases 1.0 or we just drop the dependency alltogether. Swapping it just seems like a temporary fix at this point

mstange:
[...]

PR #60, opened 2024-05-10
remove derive_more

xd009642:
First appologies for the size of this PR I can break it up if desired into a PR per file or something less giant :sweat_smile: 

This PR removes derives_more and thus the dependency on syn/quote etc. Both versions of syn take approximately 25-30s for a clean build on my system and a lot of crates currently have syn 1.x and 2.x in the dependency tree. Because of syn 2.x being blocked on a 1.0.0 release of derive_more (which seems blocked for other reasons) I suggested removing derive_more and replacing with the handrolled implementations oin #55 .

Initially I started doing this in `src/frame.rs` manually until I realised just how many impls there were to replace. After that I took the approach of using `cargo expand` and cleaning up the names and impled code to look less generated.

One thing changed - all the display impls previously had `#[inline]` as a hint I didn't persist that but I could add it in just to ensure performance doesn't change.

Ignoring dev-dependencies for a user of this project the dependency tree would look like:

```
ruzstd v0.6.0 (/home/daniel/personal/zstd-rs)
├── byteorder v1.5.0
└── twox-hash v1.6.3
    ├── cfg-if v1.0.0
    └── static_assertions v1.1.0
```

Whereas before it was:

```
ruzstd v0.6.0 (/home/daniel/personal/zstd-rs)
├── byteorder v1.5.0
├── derive_more v0.99.17 (proc-macro)
│   ├── proc-macro2 v1.0.82
│   │   └── unicode-ident v1.0.12
│   ├── quote v1.0.36
│   │   └── proc-macro2 v1.0.82 (*)
│   └── syn v1.0.109
│       ├── proc-macro2 v1.0.82 (*)
│       ├── quote v1.0.36 (*)
│       └── unicode-ident v1.0.12
└── twox-hash v1.6.3
    ├── cfg-if v1.0.0
    └── static_assertions v1.1.0
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,6 @@ categories = ["compression"]
 [dependencies]
 byteorder = { version = "1.5", default-features = false }
 twox-hash = { version = "1.6", default-features = false, optional = true }
-derive_more = { version = "0.99", default-features = false, features = ["display", "from"] }
 
 [dev-dependencies]
 criterion = "0.5"
@@ -24,7 +23,7 @@ rand = { version = "0.8.5", features = ["small_rng"] }
 [features]
 default = ["hash", "std"]
 hash = ["dep:twox-hash"]
-std = ["derive_more/error"]
+std = []
 
 [[bench]]
 name = "reversedbitreader_bench"
--- a/src/blocks/literals_section.rs
+++ b/src/blocks/literals_section.rs
@@ -14,19 +14,49 @@ pub enum LiteralsSectionType {
     Treeless,
 }
 
-#[derive(Debug, derive_more::Display, derive_more::From)]
-#[cfg_attr(feature = "std", derive(derive_more::Error))]
+#[derive(Debug)]
 #[non_exhaustive]
 pub enum LiteralsSectionParseError {
-    #[display(fmt = "Illegal literalssectiontype. Is: {got}, must be in: 0, 1, 2, 3")]
     IllegalLiteralSectionType { got: u8 },
-    #[display(fmt = "{_0:?}")]
-    #[from]
     GetBitsError(GetBitsError),
-    #[display(
-        fmt = "Not enough byte to parse the literals section header. Have: {have}, Need: {need}"
-    )]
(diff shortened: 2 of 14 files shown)


KillingSpark:
Thanks for the effort! I'll review it soon :)

xd009642:
There we go, clippy should now be happy!

xd009642:
And sorry about that fat fingered what was meant to be a simple change, all fixed now and I've ran cargo-hack and clippy locally and it passes as expected

KillingSpark:
Thanks a lot! :)

xd009642:
No worries :grin: . Btw any timeline for the next release?

KillingSpark:
I really wanna get my own PR resolved before that, it will increase performance by a few percent. I have a changeset locally that seems to do what I want. If this doesn't work out soon, I'll make a release. The PR I have there is purely internal changes so would be a minor version release, so not a big deal to push it back a bit.

JelteF:
Author of derive_more here. The `1.0.0-beta.6` release is really stable (it fixes many things compared to than the 0.99.x branch).

khuey:
It would be nice if you could cut a release with this PR in it. We're considering using ruzstd to provide support for zstd-compressed debug info in the backtrace crate, but the dependency on an old version of syn is a blocker to that.

JelteF:
The 1.0.0-beta.6 release of derive_more depends on the 2.0 version of syn. I believe that should work for you.

JelteF:
@khuey oh wait I misunderstood. You're asking for a release of zstd-rs not derive_more

khuey:
> @khuey oh wait I misunderstood. You're asking for a release of zstd-rs not derive_more

Yeah, that's correct.

I don't think we care whether this crate depends on syn 2 via derive-more 1.0 or whether it doesn't depend on syn at all but we don't want it to depend on syn 1.

KillingSpark:
@khuey If I don't find the time to work further on #58 this week I'll cut a release on the weekend. So either way next week there will be a release without the syn 1.0 dependency

PR #61, opened 2024-05-14
Add documentation throughout the codebase.

zleyyij:
I wanted to see how feasible it would be to implement a compressor for this crate, so I went through the code and added documentation for most things. Whether or not I implement a compressor in the future will depend on my motivation, but at the very least, I can contribute doc improvements.


KillingSpark:
Very nice thanks! I'll give it a read soon. 

About the compressor side of things: I would definitely accept a very basic compressor. I toyed with the idea of implementing an encoder that only supports RAW and RLE blocks. Even a "bad" compressor would enable interoperability between software that is expected to exchange data in zstd frames.

src/blocks/block.rs
```diff
@@ -17,9 +25,18 @@ impl core::fmt::Display for BlockType {
     }
 }
 
+/// A representation of a single block header. As well as containing a frame header,
+/// each Zstandard frame one or more blocks.
```
KillingSpark, line 29: each Zstandard frame *contains* one or more blocks.

src/blocks/block.rs
```diff
@@ -17,9 +25,18 @@ impl core::fmt::Display for BlockType {
...
 pub struct BlockHeader {
+    /// Whether this block is the last block in the frame.
+    /// It may be followed by an optional `Content_Checksum` if it is.
     pub last_block: bool,
     pub block_type: BlockType,
+    /// Ths size of the decompressed data. If the block type
```
KillingSpark, line 35: Ths -> This

src/decoding/sequence_section_decoder.rs
```diff
@@ -475,19 +488,34 @@ fn maybe_update_fse_tables(
     Ok(bytes_read)
 }
 
+// The default Literal Length Length decoding table uses an accuracy logarithm of 6 bits.
```
KillingSpark, line 491: Length doubled

src/frame.rs
```diff
@@ -178,10 +249,12 @@ impl FrameHeader {
         }
     }
 
+    /// Read the dictionary ID required to decode the frame from the header.
```
KillingSpark, line 252: Just "the dictionary ID required to decode the frame from the header.
Comment"?

src/fse/fse_decoder.rs
```diff
@@ -125,13 +147,20 @@ impl From<GetBitsError> for FSEDecoderError {
...
 
+/// A single entry in an FSE table.
 #[derive(Copy, Clone)]
 pub struct Entry {
+    /// This value is used as an offset value, and it is added
+    /// to a value read from the stream to determine the next state value.
     pub base_line: u32,
+    /// How many bits should be read from the stream when decoding this entry.
     pub num_bits: u8,
+    /// The byte to replace this entry with during decoding (I think?).
```
KillingSpark, line 158: /// The byte that should be put in the decode output when encountering this state

src/huff0/huff0_decoder.rs
```diff
@@ -236,10 +264,13 @@ impl<'t> HuffmanDecoder<'t> {
         Ok(num_bits)
     }
 
+    /// Advance the internal cursor to the next symbol. After this, you can call `decode_symbol`
+    /// to read from the new position.
     pub fn next_state(
         &mut self,
         br: &mut BitReaderReversed<'_>,
     ) -> Result<u8, HuffmanDecoderError> {
+        // I have no clue how this works
```
KillingSpark, line 273: This comment should be removed ;)

KillingSpark: self.state stores a window of the bit stream. The table can be indexed via this state, telling you how many bits identify the current symbol. These bits that identify the current symbol get shifted out, new bits are read from the stream and appended at the end of the current stream.

src/huff0/mod.rs
```diff
@@ -1,2 +1,6 @@
+/// Huffman coding is a method of encoding where symbols are assigned a code,
+/// and more commonly used symbols get shorter codes, and less commonly
+/// used symbols get longer codes. Codes prefix free, meaning no two codes
```
KillingSpark, line 3: Codes *are* prefix free

KillingSpark:
Hi just a few things to fix. Sorry for not making it one review I probably spammed your inbox a bit :/ 

I want to thank you again, that's a lot of good doc!

src/frame.rs
zleyyij: I reworded it in a way that's more concise, let me know if you'd prefer your wording.

src/huff0/huff0_decoder.rs
zleyyij: Documentation updated, probably needs a second set of eyes.

zleyyij:
All requested changes have been applied, there are a few things I think you should double check, but I've done everything you asked

src/huff0/huff0_decoder.rs
KillingSpark: Looks good :)

KillingSpark:
Looks good to me, the two CI errors should be easy fixes.

The doc comment should be fixable by not making it a file but just creating a reader with a `todo!()`

```
let reader: &mut Reader = todo!("Get a reader from somewhere, like a file");
```

KillingSpark:
Thanks!

PR #62, opened 2024-05-19
Implement accessors for inner reader on StreamDecoder

ifd3f:
I have an application where I access the inner Read object to get some statistics on how far we've read. This adds `get_ref` and `get_mut`, and another one where you can destructure to get both the Read and FrameDecoder.

### DIFF
--- a/src/streaming_decoder.rs
+++ b/src/streaming_decoder.rs
@@ -39,8 +39,39 @@ impl<READ: Read> StreamingDecoder<READ, FrameDecoder> {
         decoder.init(&mut source)?;
         Ok(StreamingDecoder { decoder, source })
     }
+}
 
-    pub fn inner(self) -> FrameDecoder {
+impl<READ: Read, DEC: BorrowMut<FrameDecoder>> StreamingDecoder<READ, DEC> {
+    /// Gets a reference to the underlying reader.
+    pub fn get_ref(&self) -> &READ {
+        &self.source
+    }
+
+    /// Gets a mutable reference to the underlying reader.
+    ///
+    /// It is inadvisable to directly read from the underlying reader.
+    pub fn get_mut(&mut self) -> &mut READ {
+        &mut self.source
+    }
+
+    /// Destructures this object into the inner reader.
+    pub fn into_inner(self) -> READ
+    where
+        READ: Sized,
+    {
+        self.source
+    }
+
+    /// Destructures this object into both the inner reader and [FrameDecoder].
+    pub fn into_parts(self) -> (READ, DEC)
+    where
+        READ: Sized,
+    {
+        (self.source, self.decoder)
+    }
+
+    /// Destructures this object into the inner [FrameDecoder].
+    pub fn into_frame_decoder(self) -> DEC {
         self.decoder
     }
 }
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -331,9 +331,11 @@ fn test_streaming() {
     // Test resetting to a new file while keeping the old decoder
 
     let mut content = fs::File::open("./decodecorpus_files/z000068.zst").unwrap();
-    let mut stream =
-        crate::streaming_decoder::StreamingDecoder::new_with_decoder(&mut content, stream.inner())
-            .unwrap();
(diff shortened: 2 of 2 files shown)


src/streaming_decoder.rs
```diff
@@ -39,10 +39,35 @@ impl<READ: Read> StreamingDecoder<READ, FrameDecoder> {
         decoder.init(&mut source)?;
         Ok(StreamingDecoder { decoder, source })
     }
+}
 
-    pub fn inner(self) -> FrameDecoder {
+impl<READ: Read, DEC: BorrowMut<FrameDecoder>> StreamingDecoder<READ, DEC> {
+    /// Destructures this object into the inner [FrameDecoder].
+    pub fn inner(self) -> DEC {
```
KillingSpark, line 46: I think this should be called `into_inner` for consistency with `into_inner_parts`

src/streaming_decoder.rs
```diff
@@ -39,10 +39,35 @@ impl<READ: Read> StreamingDecoder<READ, FrameDecoder> {
...
         self.decoder
     }
+
+    /// Gets a reference to the underlying reader.
+    /// 
+    /// It is inadvisable to directly read from the underlying reader.
```
KillingSpark, line 52: I think the advise is only necessary on the `get_mut`, since reading requires a mutable reference to the reader.

src/streaming_decoder.rs
```diff
@@ -39,10 +39,35 @@ impl<READ: Read> StreamingDecoder<READ, FrameDecoder> {
...
+    pub fn get_ref(&self) -> &READ {
```
KillingSpark, line 53: I think this should be called `inner` instead of the current `inner` function

src/streaming_decoder.rs
```diff
@@ -39,10 +39,35 @@ impl<READ: Read> StreamingDecoder<READ, FrameDecoder> {
...
+        &self.source
+    }
+
+    /// Gets a mutable reference to the underlying reader.
+    ///
+    /// It is inadvisable to directly read from the underlying reader.
+    pub fn get_mut(&mut self) -> &mut READ {
```
KillingSpark, line 60: And this could then be called `inner_mut`

KillingSpark (commented):
This seems like a good addition to the api of the streaming decoder. I have some thoughts about the naming, but otherwise this looks good 👍

src/streaming_decoder.rs
ifd3f: The naming of `get_mut` and `get_ref` were largely to keep consistency with other things in the ecosystem:
- [bzip2](https://docs.rs/bzip2/latest/bzip2/read/struct.BzDecoder.html#method.get_ref)
- [flate2](https://docs.rs/flate2/latest/flate2/read/struct.GzDecoder.html#method.get_ref)
- [xz2](https://docs.rs/xz/latest/xz/bufread/struct.XzDecoder.html#method.get_ref)
- [std::io::BufReader](https://docs.rs/xz/latest/xz/bufread/struct.XzDecoder.html#method.get_ref)

I think I avoided making a `into_inner() -> READ` because it would be confusing with `inner()`. But, if we're willing to change the API, we could have that as well?

src/streaming_decoder.rs
ifd3f: See comment on get_ref

src/streaming_decoder.rs
ifd3f: Maybe into_frame_decoder?

KillingSpark: I like that :+1:

src/streaming_decoder.rs
KillingSpark: I didn't realize this was a common api. Then let's keep it this way!

ifd3f:
Seems I forgot to run with --no-default-features, oops.

KillingSpark:
> Seems I forgot to run with --no-default-features, oops.

No worries, I do that all the time. But that's what the CI is here for :)

KillingSpark:
Don't worry about the clippy thing, I can fix that after merging. I'll do that later today :+1:

Thanks for the contribution!

PR #63, opened 2024-06-15
Revert "remove derive_more (#60)"

JelteF:
This reverts commit 5265c12c8cdc03647b90ae89fd3b64015a1c2d8c.

The just released `derive_more` v0.99.18 uses syn 2.x instead of syn 1.x,
removing the conflict with other dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ categories = ["compression"]
 [dependencies]
 byteorder = { version = "1.5", default-features = false }
 twox-hash = { version = "1.6", default-features = false, optional = true }
+derive_more = { version = "0.99", default-features = false, features = ["display", "from"] }
 
 [dev-dependencies]
 criterion = "0.5"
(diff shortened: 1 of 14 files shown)


KillingSpark:
Sorry for the late response!

I will have to think about this a bit. derive_more is definitely useful for developing crates, it provides a lot of developer convencience. But this crate is (probably/hopefully) not changing a lot in the future, so the convenience derive_more provides isn't that important for this crate. 

Having derive_more as a dependency inflates compile times and dependencies. This might not matter for many crates that already have a big dependency list, but this crates has very few dependencies. I think now that the effort has been made to remove derive_more it is worth it to keep the compile times and dependencies as minimal as possible.

Also I see the same kind of issue popping up in the future when/if a syn 3.0 is released, which can be prevented by keeping the current solution.

KillingSpark:
I think I'll leave it as is, thank you though!

PR #66, opened 2024-08-25
Remove byteorder dependency

workingjubilee:
In doing so, simplify bit_reader_reverse::refill_slow a tad.

The primary reason for doing this is to make the std/backtrace integration step with this crate easier. This way we have one less dependency to compile, and thus don't have to also "bless" byteorder with the special features libstd needs to compile a crate as part of its dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,6 @@ keywords = ["zstd", "zstandard", "decompression"]
 categories = ["compression"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


workingjubilee:
Fuzzing is seemingly not finding new issues with this variant.

KillingSpark:
I'm guessing this is obsolete with #67 containing these changes too?

Issue #75, opened 2024-11-27
Undefined behavior during decoding


StarStarJ:
Tried to use this crate as an alternative to the zstd crate, to allow tests running with miri, but it crashed due to UB:
```
test test::my_test ... error: Undefined Behavior: using uninitialized data, but this operation requires initialized memory
   --> /home/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ruzstd-0.7.2/src/decoding/ringbuffer.rs:462:30
    |
462 |             .write_unaligned(src.0.cast::<CopyType>().read_unaligned())
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ using uninitialized data, but this operation requires initialized memory
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: BACKTRACE on thread `test::my_test`:
    = note: inside `ruzstd::decoding::ringbuffer::copy_bytes_overshooting` at /home/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ruzstd-0.7.2/src/decoding/ringbuffer.rs:462:30: 462:71
    = note: inside `ruzstd::decoding::ringbuffer::RingBuffer::extend_from_within_unchecked` at /home/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ruzstd-0.7.2/src/decoding/ringbuffer.rs:307:17: 307:62
    = note: inside `ruzstd::decoding::decodebuffer::DecodeBuffer::repeat_in_chunks` at /home/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ruzstd-0.7.2/src/decoding/decodebuffer.rs:153:17: 154:72
    = note: inside `ruzstd::decoding::decodebuffer::DecodeBuffer::repeat` at /home/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ruzstd-0.7.2/src/decoding/decodebuffer.rs:108:17: 108:71
    = note: inside `ruzstd::decoding::sequence_execution::execute_sequences` at /home/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ruzstd-0.7.2/src/decoding/sequence_execution.rs:75:13: 77:65
[...]

PR #77, opened 2024-11-28
Representative fuzzing harness

Shnatsel:
Make the fuzzing harness more representative of real-world code. It can now find #75 in under a minute when fed zstd's [official fuzzing seeds](https://github.com/facebook/zstd/blob/dev/tests/fuzz/README.md).


KillingSpark:
Just to be certain, this doesn't still find UB after the fix, right?

Shnatsel:
No more UB detected by Address Sanitizer after 15 million executions.

KillingSpark:
Thanks!

PR #78, opened 2024-11-29
Replace `RingBuffer` internals with `VecDeque`

paolobarbolini:
This is a new attempt at replacing the internals of `RingBuffer` with `VecDeque`.

This is still WIP, unsafe comments need to be added and a lot of comments need to be added to document the many hacks we're doing to implement `extend_from_within` in a performant way despite `VecDeque` not exposing it's own internals.

Compared to the previous attempts I think LLVM made many advancements, combined with `VecDeque::extend` specializations, better handling of `MaybeUninit` and the idea to use `copy_bytes_overshooting` should make the performance penalty much lower.

It should also be possible, through a feature gate, to replace all use of unsafe code and `MaybeUninit` with zero initialized memory for a smaller performance penalty than going back to the original `Vec` implementation from years ago.

### DIFF
--- a/src/decoding/ringbuffer.rs
+++ b/src/decoding/ringbuffer.rs
@@ -1,269 +1,106 @@
-use alloc::alloc::{alloc, dealloc};
-use core::{alloc::Layout, ptr::NonNull, slice};
+use core::{cmp, hint::unreachable_unchecked, iter, mem::MaybeUninit, slice};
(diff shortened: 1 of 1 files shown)


KillingSpark:
Thanks for trying this out, it's important to check that having our own ringbuffer is still worth it.

A 20% performance hit would justify keeping it for the moment.

PR #80, opened 2024-12-18
Feat: add SIMD optimizations to the reverse bitreader

zleyyij:
This is currently a draft PR just to keep track of things. 

This change is effectively a complete rewrite of the reverse bitreader meant to enable SIMD intrinsic utilization.


KillingSpark:
I didn't do a deep dive on the SIMD instructions used here but going by the comments I think there is a misunderstanding of what the "reverse" in reverse_bitreader stands for.

The bits and bytes are in the correct order but values are read in reverse order, starting at the highest possible index.

```aaaBBBcc|cDDDDDDD```

Consider this bitstream where the values a-d are encoded. a, b and c each take 3 bits d takes 7 bits. The "normal" bitreader would read them in the order a,b,c,d. The reverse bitreader returns d,c,b,a.

I hope I'm wrong and just misunderstood what this change is trying to do :)

zleyyij:
You would be correct, sadly all of the work I've done so far has been
pointless because I misinterpreted the document.

From here, I can try to architect a new SIMD algorithm, or I can attempt to
optimize a different part of the code. Intel has a fascinating paper about
Huffman code processing with SIMD, and I'm sure there are other potential
areas available for rich optimization.

zleyyij:
After a lot of thought and headache, I've come to the conclusion that having a generic (reverse) `BitReader` as an abstraction for FSE and Huffman coding limits the levels of optimization possible because a read is performed, then it's processed, then another read is performed in a very serial manner, whereas heavy optimization will likely require batching or preprocessing in a way that the `BitReader` abstraction doesn't easily allow for.

I'll likely attempt to first migrate the Huffman decoder to a heavily optimized variant (plenty of prior work exists for SIMD Huffman decoding already exists), then move FSE decoding if feasible.

zleyyij:
I'm going to close this, I'm going to go back to the drawing board and come back when I have a better idea. I'm sure I can get somewhere, i just want to start fresh

PR #81, opened 2024-12-23
Optimize and simplify `BitReaderReversed`

fintelia:
This PR refactors the `BitReaderReversed` object based on the structure from the [Reading bits in far too many ways](https://fgiesen.wordpress.com/2018/02/19/reading-bits-in-far-too-many-ways-part-1/) blog series.

It retains the `get_bits` method to read bits from the stream, but also exposes lower-level `peek_bits`, `consume`, and `refill` methods. These new methods may unlock optimizations in other parts of the code by reducing the number of hard-to-predict branches on whether to refill the bit container. And a separate `peek_bits` method might be especially useful for Huffman decoding because it lets you do a table lookup with N bits and then decide to consume fewer than N of them, rather than the current strategy of intentionally reading past the end of the input.

On my machine (a Ryzen 5600X), I see a ~5% end-to-end performance improvement in decode time for enwik9 from this change.

### DIFF
--- a/src/decoding/bit_reader_reverse.rs
+++ b/src/decoding/bit_reader_reverse.rs
@@ -88,14 +88,14 @@ impl<'s> BitReaderReversed<'s> {
             self.refill();
         }
 
-        let value = self.peak_bits(n);
+        let value = self.peek_bits(n);
         self.consume(n);
         value
     }
 
     /// Get the next `n` bits from the source without consuming them.
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


KillingSpark:
This is an interesting changeset. I can't reproduce the performance gains on a `Ryzen 5 3500U` (it performs basically the same) but think I see the potential for the optimizations you mention in other regions of the library.

I'll look into this more after the holidays :)

KillingSpark:
Yep looks good, thanks for the PR!

PR #84, opened 2025-03-05
Filesystem refactor

zleyyij:
This pull request attempts to clean up the module tree a bit more,  bit writers/readers have moved into an `io` module, and constants have moved into a `common` module. 

More is likely to come, but I'd rather not leave you with a mega-PR


KillingSpark:
Hi, had a quick look, I like the idea but I think that the `io_traits` should stay as `io` and the bit readers should go into something specially named. The traits in io are just re-exported from `std::io` with the std feature.

Also I don't immediately see why the ci is failing. Are there behaviour changes I'm missing?

zleyyij:
Nightly clippy is failing because of a lint added just barely (<https://rust-lang.github.io/rust-clippy/master/index.html#io_other_error>, I don't know why check is failing but I can dig into it some more.

I'll try to figure out a better name for the bit reader/writer category, rename `io_traits` back to `io`, and fix CI as best I can.

zleyyij:
I haven't made any changes to the ringbuffer, so I'm extremely confused as to why it's failing.

KillingSpark:
Yeah that seems unrelated. I'll have a look at what's going on there.

zleyyij:
I fixed the `MAX_WINDOW_SIZE` issue and documented the distinction between the two max sizes. 

As for `MAX_BLOCK_SIZE`, when I referred to the spec <https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#:~:text=Window_Size-,128%20KB,-Block_Maximum_Size%20is%20constant>, I found that *both* definitions of `MAX_BLOCK_SIZE` were incorrect. When I updated the constant to 128KB rather than 128 KiB, quite a few tests began to fail.

What do you think we should do?

I think that `MAX_BLOCK_SIZE` should conform to the spec as closely as possible, and the failing tests should be fixed.

fintelia:
I think the spec means 128 KiB when it says 128 KB: https://github.com/facebook/zstd/blob/eca205fc7849a61ab287492931a04960ac58e031/doc/educational_decoder/zstd_decompress.c#L28-L29

KillingSpark:
That might be worth reporting on the zstd repo. Cyan is very friendly and I think they'd appreciate the heads up.

Anyways, I'll review this soon and merge it if I don't see any more issues

zleyyij:
https://github.com/facebook/zstd/issues/4339

KillingSpark:
Looks good to me, thanks for the work!

PR #85, opened 2025-03-17
Round 2

zleyyij:
This is a draft PR so I can ask about `reinit_from` before making sweeping changes across the codebase.

Is there any reason `mem::replace` isn't just used everywhere?

The top version is what I'm proposing
```rust
    /// Reset `self` and update `self`'s state to mirror the provided table.
    pub fn reinit_from(&mut self, other: &Self) {
        mem::replace(self, other.clone());
    }
```

This is what's currently in place
```rust
    /// Reset `self` and update `self`'s state to mirror the provided table.
    pub fn reinit_from(&mut self, other: &Self) {
        self.reset();
        self.symbol_counter.extend_from_slice(&other.symbol_counter);
        self.symbol_probabilities
            .extend_from_slice(&other.symbol_probabilities);
        self.decode.extend_from_slice(&other.decode);
        self.accuracy_log = other.accuracy_log;
    }
```
    
    When running `cargo bench`, the two benchmark with nearly identical performance. If this change was applied across the entire codebase, it could remove potentially hundreds of lines of code. I believe it may also be marginally faster because it becomes a single copy and move, rather than multiple split up.
    
    @KillingSpark thoughts?
    
    This pull request also contains some misc yardwork including:
    - Re-organizing files so that `impl`s are next to their struct definitions
    - Organizing files by entrypoint, so the top of the file is the best place to start when trying to understand it, and you can read through the file and it follows a logical progression of ideas
    - Rephrasing some documentation to be more concise


KillingSpark:
> Is there any reason mem::replace isn't just used everywhere?

I'm guessing that I just didn't know about this when I wrote that code. But to be honest, I'm not sure why this reinit is necessary. It seems like I wanted to be able to reuse some allocated ressources but, afaict that doesn't happen. So maybe there is even more cleanup possible, by making the code that uses that function simpler.

zleyyij:
Is there any reason it's preferable to reinit here?

There's already another table that's allocated, what if we just used that table?

KillingSpark:
At least theoretically there are situations where the `other` is a precomputed default table. Just copying from that to the currently used table would mean that all allocations are reused. I'm not sure if the code is currently taking advantage of that.

zleyyij:
I'll worry about that later, but I do want to circle back to it.

@KillingSpark What do you think about consolidating/improving the internal decoder interface?

There's currently `Frame`, `FrameDecoder`, `FrameDecoderState`,  `DecodeBuffer`, and `FrameDecoderScratch`.

`Frame` has one field (`FrameDecoder`) and zero methods, `FrameDecoder` stores `FrameDecoderState` which is in an `Option` but always eventually initialized.

I also feel that the `*Scratch` interface is unintuitive, would you have any ideas for how to clarify it? (naming and function calls)

What do you feel would be the best way to refactor these structs?

KillingSpark:
I think `Frame` can be ditched. It's literally just a wrapper around the header without any extra functionality.

The FrameDecoder/FrameDecoderState where separated so that the state is only initialized (-> allocates buffers) when necessary. The size of these buffers depends on the FrameHeader, so initializing them before reading the `FrameHeader` isn't really feasible.

I'm not sure how the `*Scratch` naming can be improved. They are reusable allocated spaces that are used again and again to decode stuff, which can be reused for the next block after decoding the current block.

zleyyij:
Okay, I ditched frame, and re-organized some files (most important stuff near the top, impls below structs).

Should be ready for review :)

KillingSpark:
This LGTM right now,

> What do you think about replacing *Scratch with *Cache or *DecoderCache?

It's not really a cache as it gets overwritten on each new block. It could be renamed to `*DecodingWorkspace` but that seems pretty verbose

PR #87, opened 2025-04-09
refactor: move compress_fastest to a new file

zleyyij:
This change creates a new folder `./levels`, ideally where each compression level would have its own file.

I believe this change will help keep things simple, and make addition of more compression levels easier.


KillingSpark:
Thanks!

PR #91, opened 2025-07-21
Dictionary Creation

zleyyij:
At the time of opening it's not quite in a state where it's ready to be merged, but I want you to be able to track progress and provide feedback.

My roadmap looks like this:
- [x] Design and implement an interface for churning through a lot of files without loading them all into memory at the same time
- [x] Split text into epochs
- [x] Score each segment in an epoch using a reservoir sampler to estimate frequency size within the epoch
- [x] determine the best order to concatenate k-mers to form the dictionary (likely simple)
- [ ] Implement zstd format wrappers around the dictionary
- [x] Create user-facing methods for creating a new dictionary
- [x] Create rough benchmarks to ensure that created dictionaries are effective and creation is usably fast for enwik9
- [x] Add dict creation to the CLI

Once that's done, then I could begin improving dictionary creation in a separate PR.


zleyyij:
Consider it a draft for the time being

zleyyij:
Update: the core algorithm is complete, now all that's left is cleanup, and possibly performance/efficiency improvements in a different PR.

Bench marking on the `github-users` sample set, we achieve compression gains within a percentage of the original implementation.

```
uncompressed: 100.00% (7484607)
no dict: 34.99% compression ratio (2618872)
reference dict: 16.16% (2195672 smaller) diff compared to without dict
our dict: 16.28% (2192400) diff compared to without dict⏎
```

zleyyij:
@KillingSpark this is ready for review :)

zleyyij:
The removal of `mod tests` must have happened automatically from my IDE, that's MB.

Other feedback has been addressed

KillingSpark:
Thanks!

PR #93, opened 2025-09-30
Improve Zstd binary provided

### LINKED ISSUE #38: Improve the zstd binary provided
The zstd binary was conceived as a showcase of the library but it could be useful to mimic the cli interface of the original zstd tool as close as possible

zleyyij:
Closes #38

This PR introduces the addition of several crates to improve the ergonomics of the CLI.

The new CLI has been designed to be simple, ergonomic, and scale well as more features are added. It has a real time progress bar, along with an ETA. Once processing is complete, it presents basic stats about what happened.

If demand is present, backwards compatibility could be added under a `compat` subcommand.

Future work building on this PR might involve the addition of a `bench` subcommand, and/or a `gen-dict` subcommand.

Attached is a recording of the `decompress` command in action

[asciicast](https://asciinema.org/a/wHRkOWUdYm79Yk3FcVWEgUaXu)


KillingSpark:
> Switching to a workspace makes sense, I'd be down for that and can implement that change if needed.

That would be awesome!

> I agree that the ratio feels inverted, but it's how Facebook represents it (and I think most entities). 

I think that's just for benchmarking purposes (or rather for purposes of cool graphs). "Line go up" is visually more impressive than "Line go down". Since the C impl seem to show this as a precentage then that's the way I'd go too :)

zleyyij:
Okay, switching to a workspace has been done, and the ratio was switched to a percentage (as well as a more granular time display)

Some future discussion topics:
- Can we make the test that decodes every file in `local_corpus_files` skipped by default, or removed entirely? IMO those are more integration tests, and I find myself renaming the `local_corpus_files` directory temporarily just so I can avoid processing gigabytes of data for every feature flag combo. Alternatively, updating the `.gitignore` to ignore any file with the `local_` prefix
- What do you think about adding a task runner to the repository? (i.e `just` or `cargo xtask`). This would be done to make running CI locally easier, you could have something like `just nightly-ci` and `just check`. While you can run CI locally with `act`, or by copy pasting the commands, `act` is slower than bare metal, repeats a lot of work, and creates a 30+gb docker cache for this repo alone. Copy pasting the commands is slow and easy to miss.

KillingSpark:
I created #95 to track the issue of the `local_` tests

PR #98, opened 2025-12-29
re-include LICENSE file in ruzstd crate

decathorpe:
When the ruzstd crate was moved into a subdirectory, a symlink for the README file was created, but none for the LICENSE file, which led to ruzstd 0.8.2 being published without the project's license text.


decathorpe:
Thanks for addressing this so quickly! :)

PR #101, opened 2026-03-08
Reset checksum every frames

hbina:


KillingSpark:
Oh very good catch, thanks for the bugfix!

PR #104, opened 2026-04-29
derive debug trait for compression level

feefladder:
I wanted to use it in a registry that wants to print itself, but was blocked by this.

I think it makes sense to be able to print decoding level, Debug was used elsewhere in the code and it's a rather primitive struct, so I thought this was the easiest way in - as opposed to making an issue and everything :).


KillingSpark:
Seems like a sensible addition :)

PR #105, opened 2026-05-22
Avoid expanding incompressible fastest blocks

bsutton:
now checks that a block was reduced in size. If not we use the uncompressed block.

AI generated code changes - human reviewed.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,9 @@ This document records the changes made between versions, starting with version 0
 
 # After 0.8.3 (Current)
 
+* Avoid emitting compressed blocks when the compressed payload is not smaller
+  than the raw block.
+
 # After 0.8.2
 * Introduce the `rust-version` field
 * Fix checksum generation when repeatedly using the encoder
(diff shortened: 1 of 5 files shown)


KillingSpark:
That looks like a valuable change, thanks!

Out of interest, did you actually have this issue, or did you let the AI look at the code and find any improvement?

bsutton:
the ai found the improvement.  I'm basically comparing the c version of the library with this library, with the intent of achieving similar performance. I suggested we look at the c library for possible improvements, but ai made the actual suggestions. I have a couple of experiments running hoffman encoding with about a 6% gain and current looking at matcher improvements.  I actually don't have any deep understanding of the compression methods (beyond creating dictionaries) but I to tend to know what questions to ask the ai.

bsutton:
I did suggest a couple of minor improvements to the code it generated - but I'm not a rust programmer.

KillingSpark:
Oh thats pretty clever to tell it to compare to the original library. I'd be interested if it can pinpoint or at least hint at architectural differences that cause the performance difference.

I have my suspicions but they would take a lot of effort to implement and confirm.

bsutton:
It has identified several difference. I've have it implementing each one
and then it benchmarks compression and performance.

Overnight I told it to do a CPU profile and to optimise hotspots.

When I get out of bed I will send you the results.

bsutton:
The effort on my part had been minimal.

PR #106, opened 2026-05-28
Compression/two candidate matcher

bsutton:
## Summary

Improve `CompressionLevel::Fastest` match selection by retaining both the
oldest and newest candidate for each suffix hash bucket.

Each suffix table slot stores two one-based `NonZeroU32` indexes in an
`Option<Candidates>`, keeping the per-slot storage compact on 64-bit targets
without manual bit packing. Keeping the oldest candidate preserves
long-distance repeated-run behavior, while checking the newest candidate
improves local structured-text matches.

The matcher now keeps normal slice cursors as `usize` and converts to the
compact `u32` representation only at the `SuffixStore` boundary. Collision
checks first verify the exact 5-byte match key, and candidate iteration is
explicit to avoid extra iterator overhead in the hottest path.

## Benchmarks

Benchmarked locally with deterministic fixtures. Each result is the median of
3 measured runs after one warmup run. Rust results use `ruzstd-cli compress -l
1`; C results use `/usr/bin/zstd -q -f -1`. All generated `.zst` files were
verified by decoding with `/usr/bin/zstd -d -c`.

Base: `upstream/master` at `1f95af6` (`Avoid expanding incompressible fastest blocks (#105)`)

### Compression Size

| Fixture | Upstream bytes | This PR bytes | Size improvement | C zstd -1 bytes |
| --- | ---: | ---: | ---: | ---: |
| `decodecorpus_pack.bin` | 5,976,095 | 5,725,987 | +4.2% | 5,385,951 |
| `json_logs_32m.jsonl` | 3,392,237 | 2,605,620 | +23.2% | 1,138,701 |
| `repeated_text_32m.txt` | 31,757 | 31,757 | 0.0% | 3,116 |
| `xorshift_32m.bin` | 33,555,213 | 33,555,213 | 0.0% | 33,555,214 |

### CPU

| Fixture | Upstream CPU | This PR CPU | CPU improvement | C zstd -1 CPU |
| --- | ---: | ---: | ---: | ---: |
| `decodecorpus_pack.bin` | 0.15s | 0.14s | +6.7% | 0.01s |
| `json_logs_32m.jsonl` | 0.18s | 0.19s | -5.6% | 0.02s |
| `repeated_text_32m.txt` | 0.12s | 0.13s | -8.3% | 0.01s |
| `xorshift_32m.bin` | 0.59s | 0.58s | +1.7% | 0.02s |
| **Aggregate** | **1.04s** | **1.04s** | **0.0%** | **0.06s** |

### Peak RSS

| Fixture | Upstream RSS | This PR RSS | C zstd -1 RSS |
| --- | ---: | ---: | ---: |
| `decodecorpus_pack.bin` | 6,420 KB | 6,376 KB | 21,984 KB |
| `json_logs_32m.jsonl` | 5,784 KB | 5,576 KB | 18,660 KB |
| `repeated_text_32m.txt` | 5,580 KB | 5,552 KB | 17,796 KB |
| `xorshift_32m.bin` | 6,032 KB | 6,292 KB | 25,552 KB |

## Verification

- `cargo fmt --all --check`
- `cargo test -q -p ruzstd`
- `cargo clippy -q -p ruzstd --lib -- -D warnings`

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,8 @@
 This document records the changes made between versions, starting with version 0.5.0
 
 # After 0.8.3 (Current)
+* Improve Fastest match selection by retaining both oldest and newest hash
+  candidates for each suffix bucket and using a cheaper five-byte hash.
 
 * Avoid emitting compressed blocks when the compressed payload is not smaller
   than the raw block.
--- a/ruzstd/src/encoding/match_generator.rs
+++ b/ruzstd/src/encoding/match_generator.rs
@@ -7,7 +7,7 @@
 
 use alloc::vec::Vec;
 use core::convert::TryFrom;
-use core::num::NonZeroU64;
+use core::num::NonZeroU32;
 
 use super::CompressionLevel;
 use super::Matcher;
@@ -55,23 +55,26 @@ impl Matcher for MatchGeneratorDriver {
     }
 
     fn get_next_space(&mut self) -> Vec<u8> {
-        self.vec_pool.pop().unwrap_or_else(|| {
-            let mut space = alloc::vec![0; self.slice_size];
-            space.resize(space.capacity(), 0);
-            space
-        })
+        match self.vec_pool.pop() {
+            Some(space) => space,
+            None => {
+                let mut space = alloc::vec![0; self.slice_size];
+                space.resize(space.capacity(), 0);
+                space
+            }
+        }
     }
 
     fn get_last_space(&mut self) -> &[u8] {
-        self.match_generator.window.last().unwrap().data.as_slice()
+        self.match_generator.last_entry().data.as_slice()
     }
 
     fn commit_space(&mut self, space: Vec<u8>) {
         let vec_pool = &mut self.vec_pool;
-        let suffixes = self
-            .suffix_pool
-            .pop()
-            .unwrap_or_else(|| SuffixStore::with_capacity(space.len()));
+        let suffixes = match self.suffix_pool.pop() {
+            Some(suffixes) => suffixes,
+            None => SuffixStore::with_capacity(space.len()),
+        };
         let suffix_pool = &mut self.suffix_pool;
         self.match_generator
             .add_data(space, suffixes, |mut data, mut suffixes| {
@@ -94,12 +97,21 @@ impl Matcher for MatchGeneratorDriver {
 /// This stores the index of a suffix of a string by hashing the first few bytes of that suffix
 /// This means that collisions just overwrite and that you need to check validity after a get
 struct SuffixStore {
-    // Packed oldest/newest indexes (32bits) into one 64bit word 
-    // We need 17 bits per index to store the maximum window size of 128kb.
-    // We store indexes using one-based so Option can
-    // use NonZeroU64's niche.
-    slots: Vec<Option<NonZeroU64>>,
+    slots: Vec<Option<Candidates>>,
     len_log: u32,
+}
+
+#[derive(Copy, Clone)]
+struct Candidates {
+    // We need 17 bits per index to store the maximum window size of 128kb.
+    // We store indexes using one-based so Option can use a NonZeroU32 niche.
+    oldest: NonZeroU32,
+    newest: NonZeroU32,
+}
+
+struct CandidateIndexes {
+    oldest: usize,
+    newest: Option<usize>,
 }
 
 impl SuffixStore {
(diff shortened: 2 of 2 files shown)


KillingSpark:
This is definitely an interesting idea and the benchmarks speak for themselves. The PR in its current state isn't great though.

Firstly the test loses a lot of checks. The switch from "assert_equals" to "reconstruct" loses the checks against the expected Sequences. If the expectations change because of these new mechanisms, the new expectations should be encoded there.

Secondly, the conversion of the suffix index to the store index is unnecessary. The choice of using a u32 (so we can store two of them in the u64) instead of a usize is fine, a 4Gb limit on the window size is definitely reasonable. But then we can just make the `suffix_idx` be a u32 and not have a function that does the conversion using an unwrap.

Also, there is a questionable `step_by(2)` addition without explanation. That seems... weird? I've made a comment at the change for that.

bsutton:
Let me review your notes, I have some additional improvements pulled from
the C implementation so will package them in as well.

bsutton:
hopefully this addresses your concerns. I placed the indexes into a struct as u32s remove the step call and cleaned up some casts and unwrap calls.

bsutton:
I've been running the ai overnight with the goal of adopting all of the C implementation ideas with the aim of getting equivalent performance and compression to the C library. I've also asked to to expand the corpus of fixtures based on the C corpus to ensure the bench marks are valid (I don't have results from the expanded corpus as yet). 

This is the current bench marks:

Percent improvements compare new/current against upstream. Each compressed output is decoded with C zstd and byte-compared against the original fixture.

```text
Fixture                Upstream bytes  C bytes     New bytes   % Improvement  Upstream CPU  C CPU  New CPU  % Improvement
---------------------  --------------  ----------  ----------  -------------  ------------  -----  -------  -------------
decodecorpus_pack.bin  5,976,095       5,385,951   5,371,424   +10.1%         0.13s         0.04s  0.16s    -23.1%       
json_logs_32m.jsonl    3,392,237       1,138,701   742,727     +78.1%         0.18s         0.05s  0.11s    +38.9%       
repeated_text_32m.txt  31,757          3,116       2,874       +91.0%         0.12s         0.02s  0.00s    +100.0%      
xorshift_32m.bin       33,555,213      33,555,214  33,555,210  +0.0%          0.59s         0.06s  0.02s    +96.6%       
```

bsutton:
The ai did over 150 experiments over night.

bsutton:
just an update, I've had ai running overnight trying to optimise compression and cpu.

WE now have compression that is as good as the C code - or better but at the cost of cpu. We need to revisit the cpu stats. 

The problem is that this is likely to be a large pr.

### Level 4

| Fixture | Upstream bytes | Current retained baseline | C `zstd -4` |
| --- | ---: | ---: | ---: |
| `decodecorpus_pack.bin` | `NA` / `NA` CPU | `4,675,858` / `0.83s` CPU | `4,789,813` / `0.07s` CPU |
| `json_logs_32m.jsonl` | `NA` / `NA CPU | `602,826` / `0.21s` CPU | `1,361,274` / `0.09s` CPU |
| `repeated_text_32m.txt` | `NA` / `NA` CPU | `2,874` / `0.02s` CPU | `3,128` / `0.04s` CPU |
| `xorshift_32m.bin` | `NA` / `NA CPU | `33,555,210` / `0.06s` CPU | `33,555,214` / `0.08s` CPU |

Repeat-run note:
- `decodecorpus_pack.bin` level-4 CPU stayed in the `0.84s/0.85s` median band.
- `json_logs_32m.jsonl` level-4 CPU stayed in the `0.21s/0.22s` median band.

### Level 1

| Fixture | Upstream bytes | Current retained baseline | C `zstd -1` |
| --- | ---: | ---: | ---: |
| `decodecorpus_pack.bin` | `5,976,095 ` / `0.13s ` CPU | `5,324,267` / `0.18s` CPU | `5,385,951` / `0.05s` CPU |
| `json_logs_32m.jsonl` | ` 3,392,237 ` / ` 0.18s` CPU | `690,084` / `0.12s` CPU | `1,138,701` / `0.05s` CPU |
| `repeated_text_32m.txt` | `1,757` / `0.12s` CPU | `2,874` / `0.00s` CPU | `3,116` / `0.02s` CPU |
| `xorshift_32m.bin` | `33,555,213` / ` 0.59s ` CPU | `33,555,210` / `0.02s` CPU | `33,555,214` / `0.06s` CPU |

KillingSpark:
> The problem is that this is likely to be a large pr.

How large are talking? :)

Also: I am currently wondering about how licensing works in this case. AFAIK the C project is licensed under GPL2 or BSD in some form. I intentionally did not read too much of the code of that library to make sure this is my own work and I am free to license it in any way I'd like.

Is this still possible, when the LLM is explicitly using the C library as the basis of its work?

PR #111, opened 2026-06-23
Make the decode window size limit configurable

LucaCappelletti94:
`ruzstd` caps the decode window at a hardcoded 100 MB with no override, so it cannot decode frames declaring a larger window such as anything from `zstd --long`. This adds an opt-in to raise the cap while keeping the 100 MB default for everyone else, mirroring `ZSTD_d_windowLogMax` in `libzstd`. My use case is streaming a 2 GiB-window corpus from pure Rust without pulling in C `libzstd`.

It also makes the cap apply to the first frame of a stream, which `new` skipped while `reset` enforced it, and reshapes `FrameDecoderError::WindowSizeTooBig` to report the effective limit. That last piece is a small breaking change on the unreleased 0.8.4 that I can drop if you prefer strictly non-breaking.


ruzstd/src/decoding/frame_decoder.rs
```diff
@@ -93,9 +97,13 @@ pub enum BlockDecodingStrategy {
 }
 
 impl FrameDecoderState {
-    pub fn new(source: impl Read) -> Result<FrameDecoderState, FrameDecoderError> {
+    pub fn new(
```
zleyyij: To avoid a breaking change, we could keep the old function signature and use `DEFAULT_MAX_WINDOW_SIZE` in `new()`, and expect people to call `set_max_window_size()` if they want to update it.

This would keep the `new()` interface streamlined and sufficient for most usecases.

zleyyij (commented):
Other than semantic discussion around the signature of `FrameDecoderState::new()`, this PR seems structurally sound and a reasonable enough addition.

I'm just contributing a review, and don't have final say.

Good work!

KillingSpark:
Thanks for the work!

@zleyyij maybe I am missing something but I think the signature with a breaking change is on the non-public FrameDecoderState and not the public FrameDecoder itself. If that's the case this PR looks good to me but since you brought it up I want to make sure I did not miss a breaking change there :)

The change in the error type is fine for me, I have no issue bumping to a 0.9 version instead of the next 0.8

LucaCappelletti94:
> Thanks for the work!
> 
> @zleyyij maybe I am missing something but I think the signature with a breaking change is on the non-public FrameDecoderState and not the public FrameDecoder itself. If that's the case this PR looks good to me but since you brought it up I want to make sure I did not miss a breaking change there :)
> 
> The change in the error type is fine for me, I have no issue bumping to a 0.9 version instead of the next 0.8

The parameter is only on the private `FrameDecoderState`, the public `FrameDecoder::new()` is unchanged. I also dropped its leftover confusing `pub` so it clearly reads as internal. The only public break is the `WindowSizeTooBig` reshape.

ruzstd/src/decoding/frame_decoder.rs
LucaCappelletti94: The public `FrameDecoder::new()` already works that way. The signature in the diff is the private `FrameDecoderState`, now a plain `fn` to make that obvious.

LucaCappelletti94:
will fix

KillingSpark:
Thanks again :)

ruzstd/src/decoding/frame_decoder.rs
```diff
@@ -93,9 +97,13 @@ pub enum BlockDecodingStrategy {
...
 
 impl FrameDecoderState {
-    pub fn new(source: impl Read) -> Result<FrameDecoderState, FrameDecoderError> {
+    fn new(
+        source: impl Read,
+        max_window_size: u64,
+    ) -> Result<FrameDecoderState, FrameDecoderError> {
         let (frame, header_size) = frame::read_frame_header(source)?;
         let window_size = frame.window_size()?;
+        Self::check_window_size(window_size, max_window_size)?;
```
Eh2406, line 106: Thank you for adding this check. I think before malicious input could have triggered OOM.

PR #119, opened 2026-08-17
Fix libzstd compatibility issues with DEFAULT_MAX_WINDOW_SIZE

kpcyrd:
The Arch Linux package `rust-1:1.97.1-1-x86_64.pkg.tar.zst` fails to decompress with ruzstd, but is processed without problems by pacman with libarchive.

This was traced to a mismatch in the default max window size, libzstd uses a default limit of 128MiB, which is higher than ruzstd's 100MiB. As long as this limit is never exercised everything works well, but libzstd default settings may compress data in a way that ruzstd can not decompress anymore.

This was discovered in repro-env development, as part of the Reproducible Builds project.

Debugged with help from @stoeckmann

### DIFF
--- a/ruzstd/src/decoding/frame_decoder.rs
+++ b/ruzstd/src/decoding/frame_decoder.rs
@@ -16,10 +16,13 @@ use core::convert::TryInto;
 
 /// The default maximum window size, in bytes, that a [FrameDecoder] accepts.
 ///
-/// Defaults to 100mb to bound allocation for malformed or hostile frames. The
+/// Defaults to 128mb to bound allocation for malformed or hostile frames. The
 /// spec permits far larger windows, so raise the limit with
 /// [FrameDecoder::set_max_window_size] for input you trust.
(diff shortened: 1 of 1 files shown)


KillingSpark:
Huh interesting that this never tripped up anyone else before. Are you using this as a library or are you using the cli tool? Because the library already allows you to choose a custom limit, which would allow you to circumvent this issue until a new release is cut.

kpcyrd:
Thanks! I'm using this as a library, I've discovered `StreamingDecoder::new_with_max_window_size`, switched to it in https://github.com/kpcyrd/repro-env/commit/8fa569c30b036fcaef49846866ec723c78b4bf17 and I'm currently in the process of cutting a new release of [repro-env](https://github.com/kpcyrd/repro-env). :)

KillingSpark:
That's great :)

And of course: Thanks for notifying me about that incompatibility and providing a fix!

PR #120, opened 2026-08-18
Remove redundant conditional branches

FranciscoTGouveia:
If I am not overlooking anything, I believe these branches are actually redundant.

That is because, before `if original.len() < result.len()`, there is an ` if original.len() != result.len() {
panic!()`, meaning that there are two possible paths: either `original.len() != result.len()` and there will be a panic, or `original.len() == result.len()`, meaning that we statically know that the if-statement will evaluate to `result.len()`.

For reference, these changes were found by an optimizer developed as a research project.

CC @nunoplopes

### DIFF
--- a/ruzstd/src/tests/mod.rs
+++ b/ruzstd/src/tests/mod.rs
@@ -210,11 +210,7 @@ fn test_decode_from_to() {
     }
 
     let mut counter = 0;
-    let min = if original.len() < result.len() {
-        original.len()
-    } else {
-        result.len()
-    };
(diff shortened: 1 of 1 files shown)


KillingSpark:
Hi, so this is technically correct, but there is a reason this was written as it is.

When these tests for equal length fail it is valuable for debugging purposes to iterate both slices until they differ to see if they also differ in content of if the shorter one is just truncated.

Maybe the tests could be improved while still keeping this functionallity?

PR #122, opened 2026-08-31
Decode buffer memory optimizations

badicsalex:
ruzstd decodes the same data in far more memory than it needs, its decode buffers are almost 2x bigger.

Two independent causes.

1) `read` waits for the whole request to be collectable, and `can_collect()` is `len - window_size`, so the buffer holds the window plus the caller's entire buffer before yielding a byte. So if I read in 8kb chunks, 4MiB-window frame costs ~2MiB of ring, but if I read 2MiB chunks, it's ~8MiB of ring. Since things like `read_exact()` repeat reads, we don't **need** to fill read buffers, and surprisingly it's actually faster to return short (see below)

2) The ring then rounds to the next power of two. A decode buffer settles at the window plus one undrained block plus the block being decoded, and a window is always a power of two, so this algorithm actually allocates close to the worst possible buffer size.

Two commits: return from `read` as soon as anything is collectable, then grow as `2^k + 2 * MAX_BLOCK_SIZE` past that threshold. Neither is worth anything alone, but together the decode buffers then settle at strictly under libzstd, which allocates eagerly (this package wins on the small file + big window case)

**Measured**

Peak from a counting global allocator, timings median of 120 runs. 

5MiB in, 19.85MiB out, 4MiB window

| caller            | peak                | median           |
| ----------------- | ------------------- | ---------------- |
| `io::copy`, 8KiB  | 12.38 -> 6.87MiB    | 48.86 -> 48.63ms |
| `read_to_end`     | 64.41 -> 52.66MiB   | 50.56 -> 49.26ms |
| one 100MiB buffer | 148.41 -> 106.87MiB | 52.01 -> 48.66ms |

The patchset made decoding faster, because each growth step removed was an allocation, a copy of everything accumulated so far, and a free.

Frames below their declared window are untouched: 4KiB behind a 4MiB window still allocates 24.6KiB.

**Behavior change**

Code assuming a full read sees many short reads instead of two. This is consistent with `Read`'s contract.


badicsalex:
The code was written with AI, but I went through a bunch of iterations and checked everything with my own, carbon-based eyes.

KillingSpark:
Thanks for the PR! I'm, at first glance, not sure about the calculation of `new_cap` but the reasoning seems valid. I'll think about it more in the next days.

Edit

What's bugging me is that this will always allocate an uneven amount of bytes because the +1 is added after next_power_of_two() is called. Which was also the case before this PR but it still seems silly. We could just do `required = self.cap + amount + 1` to make sure there will be enough space for the sentinel and ask for an even amount of bytes.

Most allocators would have satisfied these requests of `2^k + 1` by allocating `2^(k + 1)`which is a waste since the ringbuffer code would only use the requested capacity essential wasting halve the requested allocation.

I'll have to do a few tests myself but I think that could be enough to fix the issue of asking for double the required amount of memory and the slack might not be necessary. But I'm not sure about that yet.

/Edit


Besides that I think it makes sense to allow the short reads. It makes the code simpler and if people really need that behavior they can always use a buffered reader / read_exact

badicsalex:
Thanks! We're using this library in an embedded(ish) use-case, so even a 50% over-allocation would hurt.
The slack in this PR is not great architecturally, because it ties to how actual processing is done on the other side of the code (i.e. if code changes, it could break silently), but I couldn't really find a solution to that, other than a discoverable comment.

(I think most of the time allocators choose 2^n because a) it's what everyone else does, b) modulo by 2^n size can be really efficient, (especially for ringbuffers and dequeues), but that doesn't really apply here.)

KillingSpark:
Seems like allocators not always return 2^k memory slices, this still improves over just using doing `(cap + amount + 1).next_power_of_two()` in memory usage reported by htop.

I hope you don't mind I didn't pull but copied the changes and tweaked it locally. I'll try to remember mentioning you in the release notes. It also required fixing some tests that are implementation specific.

If you could test with the current main and confirm that this reduces the memory usage in your system that would be really cool.

In any case thanks for bringing this up!

badicsalex:
It's alright, I don't care about attribution, as long as the actual issue is fixed :)

Testing right now.

badicsalex:
For the bigger buffers, your version works correctly, but for smaller data it regresses.

The sentinel `+1` was folded into `required` before rounding, so any capacity requirement landing exactly on a power of two rounds to the next one. It costs 1.17–1.18x for windows ≤ 256 KiB, up to 1.61x for payloads smaller than their window.
It's invisible for windows above 256 KiB with payloads that fill them, because the buffer grows there in steps of at most one block, so the requirement lands inside a rounding interval rather than exactly on its edge.

(And if someone tried to optimize it with a single `reserve(window + 2 * MAX_BLOCK_SIZE)` call, it would allocate 2x than necessary)

You can test it with something like this (assuming a `pub(crate) fn decode_buffer_capacity(&self)` on the decoder):

```
// ruzstd/tests/decode_buffer_bound.rs
use std::io::{self, Write};

#[test]
fn decode_buffer_is_sized_by_the_payload() {
    const BLOCK: usize = 128 * 1024;
    const PAYLOAD: usize = 768 * 1024;

    let mut frame = Vec::new();
    let mut encoder = zstd::Encoder::new(&mut frame, 3).unwrap();
    encoder.set_parameter(zstd::zstd_safe::CParameter::WindowLog(22)).unwrap();
    encoder.write_all(&vec![0u8; PAYLOAD]).unwrap();
    encoder.finish().unwrap();

    let mut decoder = ruzstd::decoding::StreamingDecoder::new(io::Cursor::new(&frame)).unwrap();
    let written = io::copy(&mut decoder, &mut io::sink()).unwrap();
    let capacity = decoder.decoder.decode_buffer_capacity();

    assert_eq!(written as usize, PAYLOAD);
    assert!(
        capacity <= PAYLOAD + 2 * BLOCK,
        "a {} KiB payload left the decode buffer at {} KiB",
        PAYLOAD / 1024,
        capacity / 1024
    );
}
```

The fix is to pull +1 back after the rounding.

KillingSpark:
That's true but asking the allocator for 2^k + 1 just felt wrong, because it almost certainly is wasting at least _some_ actually allocated memory. Maybe there is a middleground there somewhere?

badicsalex:
Allocating 2^n+0 or 2^n+1 always allocates the same buffer, because allocators **add a chunk header** and round up to their alignment. glibc reports an identical `malloc_usable_size` for both at literally every power of two. Asking for exactly 2^k is already wasteful: at 256 KiB glibc reserves 4080 bytes past it either way, and the +1 just uses one of them.

Examples:

| requested     | actual allocation |
| ------------- | --------------- | 
| 8 (2^3)       | 24              |
| 4096 (2^12)   | 4104            | 
| 262144 (2^18) | 266224          |

You can ask the allocator for the "actually usable" slice, but that's nightly API.

PR #123, opened 2026-09-08
fix: handle degenerate single-symbol FSE distributions

lxl66566:
NOTE: This PR contains AI generated code.

Encoding data containing a single distinct symbol panics in two places:

- `build_table_from_counts`: under `avoid_0_numbit`, weight redistribution unwraps a second maximum that doesn't exist.
- Table description writing: the zero-probability run scan reads past the end of the symbol array.

Fixes:

- Skip redistribution when the distribution has a single symbol; it keeps the full table weight and encodes at 0 bits per symbol.
- Bound the zero-run scan by the table length; the outer loop already stops once the probability sum is full.

Performance Impact: 

I think this change is not expected to impact performance.


KillingSpark:
Does that actually fix a bug? Sections with only a single symbol should never be compressed using fse compression, they should just be put in a RLE block

lxl66566:
Sorry for my mistake, this PR is intended to fix a potential panic, but it takes the wrong approach. so is the correct fix to add a Sequence RLE branch, so that the single symbol never reaches the FSE table compression?

